PR #16, opened 2016-03-11
Fluent update

Prince2k3:
Added more clean methods to help build queries, Updated the Model protocol to have more convenient  methods. Added StatementGenerator Protocol so that Fluent can have NoSql support later. Adds support for the drivers to handle sql injections by outputting parameterized query and query values as separate properties (That why the drivers can bind the values to the query.)


Prince2k3:
@timominous let me know what you think about the filters in this case

timominous:
@Prince2k3 is there a need to import `Foundation`?

Prince2k3:
No... I did removed them in another PR though

PR #41, opened 2016-05-18
Make StructuredData conform to all LiteralConvertible protocols

Evertt:
This way you can initialize it like this:

    let structuredString: StructuredData = "Hello World!"
    // structuredString = .string("Hello World!")

(To be honest, I mostly made this because it helps with the operator overloading that I'd like to add back in later O:-) )


Evertt:
On second thought, this isn't really necessary since `Filter` still uses `Value` and not `StructuredData` directly.

PR #63, opened 2016-08-04
Update paths/version for new conventions

ts:


LoganWright:
This library required some other changes, I put em in another PR.

PR #105, opened 2016-09-30
Make Limit vars public

natebird:
Updates to Limit.swift
* Change `count` and `offset` vars to be public.
* Change `init` to be public.


natebird:
This allows Drivers to override the `sql(limit)` function to support other output formats.

For instance, PostgreSQL requires the output format to be `LIMIT 1 OFFSET 0` instead of `LIMIT 0, 1`.

PR #108, opened 2016-10-07
Added test for pivot database setting

sroebert:
This test shows that setting the database on the Pivot entity and swapping the generic entities, does not result in the same database.


sroebert:
I'm creating a new pull request with the same commit from a new branch.

PR #111, opened 2016-10-09
Fix missing action on pre-existing tables during preparation

cwoloszynski:
When a table already exists (so prepare is not called), the model.database is not set.  This corrects that issue.

### DIFF
--- a/Sources/Fluent/Preparation/Database+Preparation.swift
+++ b/Sources/Fluent/Preparation/Database+Preparation.swift
@@ -11,6 +11,11 @@ extension Database {
         do {
             // check to see if this preparation has already run
             if let _ = try Migration.query().filter("name", preparation.name).first() {
+                
+                // set the current database on involved Models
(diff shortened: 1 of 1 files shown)


cwoloszynski:
@natebird I'm new to all of this; what do you mean by a test deviation?  Sorry for being a newbie!

natebird:
@cwoloszynski No problem. I'm glad you are participating in the project.

If you look at the existing [tests] (https://github.com/vapor/fluent/blob/master/Tests/FluentTests/PreparationTests.swift) that cover this part of the code, you'll see that there are a number of 'deviations' that are tested in the same long test. Your code should have it's own test by the test for the code you are modifying.

Let me know if you have any more questions.

natebird:
This fixes #110

natebird:
This PR is the same as #118 which has now been merged and released as [1.0.5](https://github.com/vapor/fluent/releases/tag/1.0.5).

PR #112, opened 2016-10-10
Add datetime support in Fluent and in Node

cwoloszynski:
I am submitting a series of pull requests to add Date as a class in Node and datetime and timestamp support in Fluent.  This allows for full support for time throughout Vapor.

Specifically, there are pull requests for Vapor, Fluent, Node, JSON, MySQL, Leaf, and FluentMySQL

### DIFF
--- a/Sources/Fluent/Preparation/Database+Preparation.swift
+++ b/Sources/Fluent/Preparation/Database+Preparation.swift
@@ -11,6 +11,11 @@ extension Database {
         do {
             // check to see if this preparation has already run
             if let _ = try Migration.query().filter("name", preparation.name).first() {
+                
+                // set the current database on involved Models
+                if let model = preparation as? Entity.Type {
+                    model.database = self
+                }
                 return true
             }
         } catch {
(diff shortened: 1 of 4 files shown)


natebird:
I'm for adding more database field types to fluent but I think we should come up with a cohesive strategy for this since it affects so many different data stores. In PostgreSQL `datetime` could be the same as `timestamp` for instance so this change would cause confusion as to which one to use.

I'd like to see more discussion about this. It seems like we could have `date`, `time`, and `timestamp` with a timezone option. Any other suggestions?

I have a issue open (https://github.com/vapor/postgresql-driver/issues/9) about adding these as well as additional native types (JSON, UUID) to the Postgresql stack.

cwoloszynski:
I'm game to a richer set of options over time. I added support in Fluent for both SQL types timestamp and datetime as one type in Node, Date. 

I believe that time zones are really a formatter issue and not a native type, at least in my head. I'd leave that out of Fluent and just let folks apply a timezone during formatting. 

Charlie

LoganWright:
Part of our deferred group for 2.0 discussion.

PR #144, opened 2016-12-27
Allow individual entities to override database driver's default 'idKey'

LuizZak:
This is my very first pull request on GitHub, so advanced apologies if I'm missing a step or two.

I'd appreciate some feedback on this change, which aims to allow implementers of 'Entity' protocol to override the `Driver.idKey` property individually, by implementing a non-nil response to the `Entity.idKey` static property. Specially with legacy databases, the identity column for the table might not always be named `id`; this change would allow such tables to be mapped into Fluent entities.

This change also affects queries generated by unions and pivots.

I'm thinking a better strategy would be to centralize a method to fetch the `idKey`, instead of manually doing it everywhere with nil-coalescings, but right now I'm not sure exactly where to refactor that to.

Relevant issue: #140


Sources/Fluent/Entity/Entity.swift
```diff
@@ -19,6 +19,15 @@ public protocol Entity: Preparation, NodeConvertible {
...
     static var name: String { get }
+    
+    /**
+        The name of the column that corresponds
+        to this entity's key.
+     
+        Return nil to use the database's global
+        idKey.
+     */
+    static var idKey: String? { get }
```
LoganWright: I'm wondering if this should be `String` and just return the `database.driver.idKey` by default so that we can deal with this everywhere internally w/o user additions.

LuizZak: In the beginning I had a rationale for that choice, but forgot to note it down, and I guess it became invalid during implementation, anyways. I'll switch it to a default and then we'll see how it looks.

LuizZak: Trying to re-implement I now seem to remember why: If let non-nil (with a default `database?.driver.idKey ?? "id"`), if the entity is used in another database without changing the static `database` variable, it'd reference the first database's idKey value in the second database instead, so the behavior would change in that case.

LuizZak:
@LoganWright I've tweaked the nullability, can you take a look at how it seems now?

Sources/Fluent/Entity/Entity.swift
```diff
@@ -138,11 +152,11 @@ extension Entity {
         Finds the entity with the given `id`.
     */
     public static func find(_ id: NodeRepresentable) throws -> Self? {
-        guard let idKey = database?.driver.idKey else {
+        if(database == nil) {
```
LoganWright: Let's change this to `guard let _ = database else { return nil }` (in general, avoid parens to stick w/ our style 👍 )

Sources/Fluent/Memory/MemoryDriver.swift
```diff
@@ -79,7 +79,7 @@ public final class MemoryConnection: Connection {
             guard let data = query.data else {
                 throw MemoryDriver.Error.dataRequired
             }
-            let i = group.create(data, idKey: driver.idKey)
+            let i = group.create(data, idKey: T.idKey)
```
LoganWright: 👍

Sources/Fluent/Relations/Children.swift
```diff
@@ -13,10 +13,8 @@ extension Children: QueryRepresentable {
         guard let ident = parent.id else {
             throw RelationError.noIdentifier
         }
-
-        let query = try T.query()
-
-        let foreignId = foreignKey ?? "\(type(of: parent).name)_\(query.idKey)"
+        
+        let foreignId = foreignKey ?? "\(type(of: parent).name)_\(T.idKey)"
```
LoganWright: While we're in here, this reads a little weird to me now that we've clarified the latter.

`"\(T.name)_\(T.idKey)"` would be more understandable I think.

LoganWright:
@LuizZak I brought your branch into our repo, this is being closed for #175

LuizZak:
Cool! Glad to contribute.

Issue #143, opened 2016-12-25
Missing NULL filter  fx .filter("column", .isNull) & .filter("column", .isNotNull)


Casperhr:
Querying database for NULL values seems impossible with Vapor. Need to use raw querying to achieve this. A query comparison filter .isNull could be an elegant solution (as well as an .isNotNull filter)

Casperhr:
Model.query().raw("WHERE BLA BLA").first()

the raw("") is also needed

BrettRToomey:
I spoke to tanner about this a few days ago. It can be accomplished by one of the following:

```swift
.filter("field", Node.null)
.filter("field", .equals, Node.null)
.filter("field", .notEquals, Node.null)
```

According to Tanner, `nil` should work, but the compiler was having a rough time trying to disambiguate it. I may have to do a PR.

PR #145, opened 2016-12-28
Add .filter("field", .isNull) and .filter("field", .isNotNull) support 

LuizZak:
Currently, it's possible to null-check fields with `.filter("field", .equals/.notEquals, Node.null)` which has a special case handling in the General SQL serializer. This adds support for proper nullability checks with a new `Nullability` enum in the Filter struct, which then makes possible to express null checks in a more explicit manner using `.filter("field", .isNull)` and `.filter("field", .isNotNull)`.

Feedback would be appreciated!

Relevant issue: #143

### DIFF
--- a/Sources/Fluent/Memory/Memory+Filters.swift
+++ b/Sources/Fluent/Memory/Memory+Filters.swift
@@ -55,6 +55,13 @@ extension Node {
 
     func passes(_ filter: Filter) -> Bool {
         switch filter.method {
+        case .nullability(let key, let nullability):
+            switch(nullability) {
+            case .isNull:
+                return self[key] == nil
+            case .isNotNull:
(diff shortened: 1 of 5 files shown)


LoganWright:
@LuizZak realized we never followed up on this. The reason you're sitting in PR purgatory is because this PR falls into a category we've become all to familiar with ... breaking changes.

Unfortunately, as great as Enums are in Swift, they kind of kill our ability to update APIs without incrementing major versions. This will have to come up with an alternative solution or be revisited for 2.0 :(

Casperhr:
@LoganWright maybe consider just making some filterNull and filterNotNull?

LoganWright:
The underlying motivation for this has been fixed to allow `nil` 👍

PR #148, opened 2016-12-30
Callbacks fix to prevent data loss

raulriera:
## Problem

Fixes issue https://github.com/vapor/vapor/issues/776

## Solution

Implemented the proposal in the previously mentioned ticket


raulriera:
Hmmmm I added a test, how could it be decreased? 🤔

PR #191, opened 2017-02-20
Child relation fix

0xTim:
This PR fixes getting children from the database when using the default Foreign ID caused by a bug where it was assuming the Parent was the same type as the Child

Unsure about the `lowercased()` call - it needs to be done somewhere, but not sure if this is the correct place or not.

Also I've chosen 138 to merge into, but we probably need a 1.4 branch to put these changes onto for a release as Fluent 2 has changed completely (though the tests may be useful)

### DIFF
--- a/Sources/Fluent/Relations/Children.swift
+++ b/Sources/Fluent/Relations/Children.swift
@@ -14,7 +14,7 @@ extension Children: QueryRepresentable {
             throw RelationError.noIdentifier
         }
         
-        let foreignId = foreignKey ?? "\(type(of: parent))_\(T.idKey)".lowercased()
+        let foreignId = foreignKey ?? "\(type(of: parent).name)_\(T.idKey)".lowercased()
         return try T.query().filter(foreignId, ident)
     }
 }
--- a/Sources/Fluent/Relations/Children.swift
+++ b/Sources/Fluent/Relations/Children.swift
@@ -14,7 +14,7 @@ extension Children: QueryRepresentable {
             throw RelationError.noIdentifier
         }
         
-        let foreignId = foreignKey ?? "\(T.name)_\(T.idKey)"
+        let foreignId = foreignKey ?? "\(type(of: parent))_\(T.idKey)".lowercased()
         return try T.query().filter(foreignId, ident)
     }
 }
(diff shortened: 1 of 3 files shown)


Sources/Fluent/Relations/Children.swift
```diff
@@ -14,7 +14,7 @@ extension Children: QueryRepresentable {
             throw RelationError.noIdentifier
         }
         
-        let foreignId = foreignKey ?? "\(T.name)_\(T.idKey)"
+        let foreignId = foreignKey ?? "\(type(of: parent))_\(T.idKey)".lowercased()
```
LoganWright: I don't think this is correct, were you seeing issues w/ `T.name`? I'm wondering if it might be good to have as `foreignKey ?? T.foreignIdKey` as well.

Tests/FluentTests/ChildTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        self.name = name
+    }
+    
+    required init(node: Node, in context: Context) throws {
+        id = try node.extract("id")
+        name = try node.extract("name")
+    }
+    
+    static func prepare(_ database: Fluent.Database) throws {
+        try database.create("Owners") { pet in
```
LoganWright: lets try to always use `create(entity)` users get these two mixed up so we're trying to make internal code consistent.

Tests/FluentTests/ChildTests.swift
```diff
@@ -0,0 +1,112 @@
...
+            pet.id()
+            pet.string("name")
+        }
+    }
+    static func revert(_ database: Fluent.Database) throws {
+        try database.delete("Pets")
```
LoganWright: `delete(Pet.entity)`

Tests/FluentTests/ChildTests.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+    
+    required init(node: Node, in context: Context) throws {
+        id = try node.extract("id")
+        name = try node.extract("name")
+        ownerId = try node.extract("owner_id")
+    }
+    
+    static func prepare(_ database: Fluent.Database) throws {
+        try database.create("Pets") { pet in
```
LoganWright: `.create(entity)`

Sources/Fluent/Relations/Children.swift
0xTim: Looks like `foreignIdKey` isn't in Fluent 1, so have used name for now

0xTim:
Changes made

0xTim:
Are there any more changes needed for this PR?

PR #226, opened 2017-04-06
Correction to EntityError's suggested fix

bygri:
For my entity `MyEntity`, I was suggested to call `database.prepare(MyEntity.self)`.

However this is not valid, and Fluent's tests use the format `MyEntity.prepare(database)` instead.

This PR changes the suggestion text to the latter format.


bygri:
Actually, both are valid if `MyEntity` conforms to `Preparation` as it ought to. Never mind!

PR #240, opened 2017-04-26
Fix index and foreign keys ignoring the escape function

vzsg:
The backtick was hardcoded in these functions, which is not compatible with Postgres. I presume this is a leftover from a previous version, as table and column names are all dynamically escaped.

With this change, index and foreign key definitions will be too.


natebird:
👍

vzsg:
No other changes are needed, `escape` is already an open function. See also: your fork of the postgresql-driver repo, I couldn't sleep so I pushed it forward a little.

natebird:
OK. Cool. I did see it in the postgresql-driver repo. Thanks! Now, get some sleep. 😴

PR #303, opened 2017-08-12
[Swift 4] 'Scope' is not a member type of 'Filter'

mozharovsky:
## Description 

Building `fluent` with Swift 4 PM fails with the following error: 

```shell
- $ swift build
- $ /Sources/Fluent/Query/Filter/NodeUtilities/Scope+String.swift:2:18: error: 'Scope' is not a member type of 'Filter' ... 
```

This looks weird, especially considering the fact that moving the `Scope` extension into `Filter.swift` file or simply creating a new file (e.g. `Filter+Scope.swift` which in this case looks good) would fix the problem. (exactly what has been done in this PR). 

Also, there are various redundant conformance constraint warnings removed with these changes: 

`... warning: redundant conformance constraint 'E': 'Entity'`

--
Update 1: CircleCI would not configure for some reason on my own. 

--
Update 2: Got CI working, all tests have been passed.


mozharovsky:
Thank you for reviews @BrettRToomey, @tanner0101!

Just committed requested changes, not sure though if `SQLiteSerializer.swift` file was supposed to be in the 755 mode. 🤔

Looking forward to further updated! 🚀

PR #360, opened 2018-01-22
Changed query comparison operators paths to KeyPath

calebkleveter:
This PR modifies the query comparison operator argument types:

```swift
public func == <Model, Value>(lhs: KeyPath<Model, Value>, rhs: Value) -> ModelFilterMethod<Model>
public func == <Model, Value>(lhs: KeyPath<Model, Value?>, rhs: Value) -> ModelFilterMethod<Model>
public func != <Model, Value>(lhs: KeyPath<Model, Value>, rhs: Value) -> ModelFilterMethod<Model>
```

By changing `ReferenceWritableKeyPath` for `KeyPath`, we can compare constant, as well as variable model properties. Previously, trying to use the operator with a constant would result in an error something like this:

    Binary operator '==' cannot be applied to operands of type 'KeyPath<Model, String>' and 'String'


twof:
What was the reason for using `ReferenceWritableKeyPath` in the first place?

calebkleveter:
No idea. @tanner0101 was open to the idea of changing it when we discussed it on Slack, so I decided a PR would be nice.

PR #400, opened 2018-03-08
Add a test to ensure that `AND` queries actually do an `AND` query

MrMage:
... and not an `OR` query. See https://github.com/vapor/fluent/pull/399 for context.


MrMage:
Sorry, those tests should be in that PR (#399) of course (I thought the impl and tests were in different repos). Closing.

PR #434, opened 2018-04-04
QueryDataDecoder should be public

rafiki270:
for all the tinkerers out there
CC: @calebkleveter


rafiki270:
this does not need to be

PR #445, opened 2018-04-11
Fix Timestampable when the models are structs.

### LINKED ISSUE #444: Timestampable doesn't work with structs.
The timestamps are lost in lines like [this one](https://github.com/vapor/fluent/blob/master/Sources/Fluent/Query/Builder/QueryBuilder+CRUD.swift#L26) because the copy of the struct with the dates is not the one assigned to `copy`.

ccrazy88:
Fixes #444.

Use the same strategy, `as!`, that is used for `SoftDeletable`.

If there are tests that I can/should add to a different repository, please let me know.


ccrazy88:
Oh, I found the FluentBenchmark folder. I will add tests.

PR #447, opened 2018-04-12
Expanded `find` function on Model

damirstuhec:


damirstuhec:
@tanner0101 People were pretty excited about this when I presented it in Slack. But I understand if this doesn't get it.

PR #579, opened 2018-10-29
Return updated/created/saved model from Future

zdnk:
Given this code where `Client` is `Model` and `struct`:
```
        return try request.fromInput(to: Client.self)
            .save(on: request)
            .map {
                return try Client.Output(
                            id: $0.requireID(),
                            platform: $0.platform,
                            role: $0.role,
                            campaignId: $0.campaignId,
                            lastModifiedAt: $0.lastModifiedAt()
                )
            }
            .encode(status: .created, for: request)
```

The code will fail on `requireID()` because current implementation of `Future.save(on:)` returns the model that was passed to it, meaning the ID is still nil.

This PR fixes this scenario and return the return value from corresponding CRUD operations on `Model` like `save(on:)` etc.

### DIFF
--- a/Sources/Fluent/Model/Model+CRUD.swift
+++ b/Sources/Fluent/Model/Model+CRUD.swift
@@ -70,21 +70,21 @@ extension Future where T: Model {
     /// See `Model`.
     public func save(on connectable: DatabaseConnectable) -> Future<T> {
         return self.flatMap(to: T.self) { (model) in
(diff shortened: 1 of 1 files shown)


0xTim:
This definitely looks like a bug/oversight

penny-coin:
Hey @zdnk, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #596, opened 2018-12-02
Aggregate with custom alias identifier

rafiki270:


Sources/FluentSQL/SQL+QuerySupporting.swift
```diff
@@ -158,7 +158,7 @@ extension QuerySupporting where
     QueryKey.Expression == QueryKey.Expression.Function.Argument.Expression
 {
     /// See `QuerySupporting`.
-    public static func queryAggregate(_ name: QueryAggregate, _ fields: [QueryKey]) -> QueryKey {
+    public static func queryAggregate(_ name: QueryAggregate, _ fields: [QueryKey], _ identifier: String) -> QueryKey {
```
rafiki270: reason I created a new method is that default values can not be protocolised in the future, happy to change if needed

PR #600, opened 2018-12-13
Add Distinct to FluentSQLQuery

bygri:
References #599.

This modifies protocol `FluentSQLQuery` to allow the MySQL and PostgreSQL drivers to implement `SELECT DISTINCT`.

This is a breaking change that requires updates to everything that depends on FluentSQLQuery. I have submitted PRs to [fluent-mysql](https://github.com/vapor/fluent-mysql/pull/135), [fluent-postgresql](https://github.com/vapor/fluent-postgresql/pull/102) and [fluent-sqlite](https://github.com/vapor/fluent-sqlite/pull/27).

NB: This doesn't enable this usage: `MyObject.query(on: worker).distinct().all()`. Instead:

```swift
let builder = MyObject.query(on: worker)
builder.query.distinct = .distinct
return builder.all()
```

### DIFF
--- a/Sources/FluentSQL/FluentSQLQuery.swift
+++ b/Sources/FluentSQL/FluentSQLQuery.swift
@@ -1,13 +1,15 @@
 public protocol FluentSQLQuery {
     associatedtype Statement: FluentSQLQueryStatement
+    associatedtype Distinct: SQLDistinct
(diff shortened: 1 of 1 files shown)


valeriomazzeo:
I think this definition should be added in `QuerySupporting` as well.

PR #628, opened 2019-05-02
Issue627: Delete of soft deleted rows fails

Jerry-Carter:
Simple fix and associated unit test.

### DIFF
--- a/Sources/Fluent/Model/Model+CRUD.swift
+++ b/Sources/Fluent/Model/Model+CRUD.swift
@@ -49,7 +49,7 @@ extension Model where Database: QuerySupporting {
     ///     - conn: Database connection to use.
     /// - returns: Future that will be completed when the delete is done.
     public func delete(force: Bool = false, on conn: DatabaseConnectable) -> Future<Void> {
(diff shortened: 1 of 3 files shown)


Sources/Fluent/Model/Model+CRUD.swift
```diff
@@ -49,7 +49,7 @@ extension Model where Database: QuerySupporting {
     ///     - conn: Database connection to use.
     /// - returns: Future that will be completed when the delete is done.
     public func delete(force: Bool = false, on conn: DatabaseConnectable) -> Future<Void> {
-        return Self.query(on: conn).delete(self, force: force)
+        return Self.query(on: conn, withSoftDeleted: true).delete(self, force: force)
```
MrMage: I think `withSoftDeleted` should only be true if `force` is true as well.

Jerry-Carter: I agree.  I'll add a case to the unit test and update the code.

Jerry-Carter: I've updated the code and extended the unit tests to cover the deletedAt property.

PR #674, opened 2020-04-07
Fluent Session migration bugfix

fananek:
- Add public init() for `CreateSession`. This solves code protection level error when calling CreateSession() in app migration.
- Renamed Fluent Session table from `sessions` to `_fluent_sessions`. It's now obvious that `_fluent_sessions` is a system table. Since we already have `_fluent_migrations` I would continue with the same naming convention.


fananek:
Concept of Fluent Sessions has been changed so this PR don't make sense anymore. Thanks for renaming the sessions table to `_fluent_sessions`.

PR #675, opened 2020-04-12
Adds EntityTaggableModel for ETag responses

grosch:
Adds a `verifyETag(on:)` to `EventLoopFuture` to verify incoming ETag.

Adds a `ModelContent` protocol which tags `Content` which includes an `init(model:)` constructor to generate a DTO based on a model.

Adds `withETag` and `createdWithETag` methods to generate a `Response` that includes the etag and location (on created) headers.

**Must be released at the same time as https://github.com/vapor/fluent-kit/pull/243**

### DIFF
--- a/Sources/Fluent/etag/Fluent+EntityTaggableModel.swift
+++ b/Sources/Fluent/etag/Fluent+EntityTaggableModel.swift
@@ -1,0 +1,20 @@
+import Vapor
+
+public extension EventLoopFuture where Value: EntityTaggableModel {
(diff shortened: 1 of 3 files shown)


grosch:
@tanner0101 @gwynne Thoughts?

grosch:
It's only here because this is where `Model` is.  If you want to put a protocol into vapor/vapor that knows about `Model` I'm fine with it moving there.

I opened an issue here: https://github.com/vapor/vapor/issues/2318

PR #742, opened 2022-06-04
Allow explicit authentication of `ModelAuthenticatable` types

s-k:
Types conforming to `ModelAuthenticatable` can be authenticated by calling `authenticate()` on the type. For example:
```swift
let user = try await User.authenticate(username: username, password: password, for: req)
```
This logs the user having the given username and password into `req.auth` and returns the found user. If the user is not found or the wrong password is supplied, an error will be thrown.

### DIFF
--- a/Sources/Fluent/ModelAuthenticatable.swift
+++ b/Sources/Fluent/ModelAuthenticatable.swift
@@ -1,3 +1,4 @@
+import FluentKit
 import Vapor
 
 public protocol ModelAuthenticatable: Model, Authenticatable {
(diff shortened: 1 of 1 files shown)


s-k:
Is there any way I can help to get the PR reviewed? 🙂

Thanks in advance!

@gwynne @Joannis @siemensikkema @0xTim

0xTim:
I'm going to throw this out to the community for comment. I'm not 100% convinced of the need for this vs just someone adding it as an extension in their project if they log in the user in multiple places. 

(To be clear, this is useful and might get used but in general we don't want to add APIs for every possible use case as it adds more for us to maintain)

gwynne:
I agree with @0xTim - while this is useful, it's behavior is too specific to its use case to justify including it in Fluent's public API.

s-k:
I understand. Thanks for considering it.

PR #745, opened 2022-09-02
Add a --yes/-y flag for migrations

PopFlamingo:
This flag enables bypassing interactive confirmation prompts which is useful for scripts where an interactive prompt cannot be used

Example usage:
```bash
vapor run migrate -y
```


PopFlamingo:
Feature is actually already globally implemented in ConsoleKit

PR #756, opened 2023-02-12
Add migrationLogLevel configuration

madsodgaard:
Adds the possibility to configure the log level for migrations with:
```swift
app.fluent.migrationLogLeveL = .debug
```

### DIFF
--- a/Sources/Fluent/FluentProvider.swift
+++ b/Sources/Fluent/FluentProvider.swift
@@ -52,7 +52,7 @@ extension Application {
             databases: self.databases,
             migrations: self.migrations,
             logger: self.logger,
-            on: self.eventLoopGroup.next(),
+            on: self.eventLoopGroup.any(),
             migrationLogLevel: self.fluent.migrationLogLevel
         )
     }
--- a/Sources/Fluent/FluentProvider.swift
+++ b/Sources/Fluent/FluentProvider.swift
@@ -52,7 +52,8 @@ extension Application {
             databases: self.databases,
             migrations: self.migrations,
             logger: self.logger,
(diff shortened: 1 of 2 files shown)


Sources/Fluent/FluentProvider.swift
```diff
@@ -128,11 +131,17 @@ extension Application {
...
                 threadPool: self.application.threadPool,
-                on: self.application.eventLoopGroup
+                on: self.application.eventLoopGroup,
+                migrationLogLevel: .info
             )
             self.application.lifecycle.use(Lifecycle())
             self.application.commands.use(MigrateCommand(), as: "migrate")
         }
+        
+        public var migrationLogLevel: Logger.Level {
```
madsodgaard, line 141: @gwynne I am not sure if we should synchronize access to this property. But users will probably only change this during configuration anyhow

gwynne: There are several other properties that'd be affected if we tried to deal with it for this one - it's considered API contract that you can't modify the config once the app is booted (though I'm not sure that's explicitly stated anywhere).

Sources/Fluent/FluentProvider.swift
```diff
@@ -52,7 +52,8 @@ extension Application {
             databases: self.databases,
             migrations: self.migrations,
             logger: self.logger,
-            on: self.eventLoopGroup.next()
+            on: self.eventLoopGroup.next(),
```
gwynne, line 55:
```suggestion
            on: self.eventLoopGroup.any(),
```

gwynne (approved):
👍

PR #758, opened 2023-02-24
Add API Docs

0xTim:
- Update SPI info
- Update CI
- Update API Docs workflow
- Add DocC bundle
- Wrap exports
- Fix the imports
- Expand index

### DIFF
--- a/Sources/Fluent/Exports.swift
+++ b/Sources/Fluent/Exports.swift
@@ -2,10 +2,10 @@
 
 @_exported import FluentKit
 
-#endif
-
 infix operator ~~
 infix operator =~
 infix operator !~
 infix operator !=~
 infix operator !~=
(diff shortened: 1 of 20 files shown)


Sources/Fluent/Exports.swift
```diff
@@ -1,5 +1,9 @@
+#if !BUILDING_DOCC
+
 @_exported import FluentKit
 
+#endif
+
 infix operator ~~
 infix operator =~
 infix operator !~
```
gwynne, line 9: These are technically re-export declarations as well and should also be within the exclusion. (Wish GH would let me select all five lines...)

gwynne (changes requested):

gwynne (approved):

PR #767, opened 2023-08-11
Various minor tidying

gwynne:
Cleans up `Exports.swift`, the README, and the CI as usual.

### DIFF
--- a/Sources/Fluent/Exports.swift
+++ b/Sources/Fluent/Exports.swift
@@ -1,11 +1,15 @@
-#if !BUILDING_DOCC
+#if swift(>=5.8)
+
+@_documentation(visibility: internal) @_exported import FluentKit
+
+#else
 
 @_exported import FluentKit
+
+#endif
 
 infix operator ~~
 infix operator =~
 infix operator !~
 infix operator !=~
 infix operator !~=
-
-#endif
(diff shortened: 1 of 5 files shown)


Sources/Fluent/Exports.swift
```diff
@@ -1,11 +1,15 @@
-#if !BUILDING_DOCC
+#if swift(>=5.8)
```
MahdiBM, line 1: Not sure how these documentation changes will play out, but i assume you already know they're fine.

MahdiBM (approved):

Sources/Fluent/Exports.swift
gwynne: The `BUILDING_DOCC` check is a relic from Tim's initial switchover of the API docs to using DocC - `@_documentation(visibility: internal)` didn't exist quite yet, so I'd suggested that we wrap the `@_exported` usages with a build condition to get around the "makes DocC build massively too many docs" problem. Naturally, `@_documentation` magically appeared less than a week later (having been originally designed pretty much specifically for this exact purpose and then generalized into having other legitimately useful uses to justify its existence), so - since we always build the docs with 5.8 (or rather, more recently, a 5.9 nightly!) the build condition was easily and much more cleanly replaced with the version check, and this is one of the last two repos which still hasn't gotten switched over (the other being `leaf`, which I'll have a PR for shortly).

PR #774, opened 2024-04-27
Fix up Fluent provider for Sendable-correct FluentKit

gwynne:
Updates the Fluent provider for the 1.48.0 release of FluentKit, which adds mostly-complete `Sendable`-correctness. Also bumps minimum Swift version to 5.8, matching FluentKit.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.8
 import PackageDescription
 
 let package = Package(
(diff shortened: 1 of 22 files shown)


Package.swift
```diff
@@ -13,18 +13,37 @@ let package = Package(
...
+
+var swiftSettings: [SwiftSetting] { [
+    .enableUpcomingFeature("ConciseMagicFile"),
+    .enableUpcomingFeature("ForwardTrailingClosures"),
+    .enableUpcomingFeature("ImportObjcForwardDeclarations"),
+    .enableUpcomingFeature("DisableOutwardActorInference"),
+    .enableUpcomingFeature("IsolatedDefaultValues"),
+    .enableUpcomingFeature("GlobalConcurrency"),
+    .enableUpcomingFeature("StrictConcurrency"),
+    .enableExperimentalFeature("StrictConcurrency=complete"),
```
MahdiBM, line 48: You did it the other way around. You should remove strict concurrency from here and add existential any, since this is for Swift 5.8.
Also you don't need to use the upcoming StrictConcurrency flag. They did upgrade the flag to an upcoming feature for Swift 5.10, but it's backward compatible.

MahdiBM (changes requested):

MahdiBM (approved):
Alright, fine. 😛

PR #775, opened 2024-05-30
Concurrency updates and database logging APIs

gwynne:
This update contains the following changes:

- Two new APIs are available: `Application.db(_:logger:)` and `Request.db(_:logger:)`. These work exactly the way the respective `.db(_:)` methods do, except the database's logger will be set to the one provided rather than the default. These APIs will hopefully make it easier to work around the unfortunate fact that `Database.logging(to:)` and its SQLKit counterpart basically don't work at all.
- The `--auto-migrate` and `--auto-revert` commandline flags no longer call `EventLoopFuture.wait()` when the app is booted in `async` mode (no more crash risk).
- The `migrate` command now has nicer console output.
- All tests have been updated to be fully `async`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,4 @@
 version: 2
-enable-beta-ecosystems: true
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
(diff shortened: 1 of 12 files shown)


0xTim (approved):
LGTM

PR #794, opened 2025-09-25
Update minimum Swift version

gwynne:
Now requires Swift 6, and `swift-format` was run on everything. Converted the tests to SwiftTesting, too.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.9
+// swift-tools-version:6.0
 import PackageDescription
 
 let package = Package(
@@ -13,8 +13,8 @@ let package = Package(
         .library(name: "Fluent", targets: ["Fluent"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/vapor/fluent-kit.git", from: "1.49.0"),
-        .package(url: "https://github.com/vapor/vapor.git", from: "4.101.0"),
+        .package(url: "https://github.com/vapor/fluent-kit.git", from: "1.52.2"),
+        .package(url: "https://github.com/vapor/vapor.git", from: "4.117.0"),
     ],
     targets: [
         .target(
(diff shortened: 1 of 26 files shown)


Package.swift
```diff
@@ -39,9 +39,10 @@ let package = Package(
...
-    .enableUpcomingFeature("ConciseMagicFile"),
-    .enableUpcomingFeature("ForwardTrailingClosures"),
-    .enableUpcomingFeature("ImportObjcForwardDeclarations"),
-    .enableUpcomingFeature("DisableOutwardActorInference"),
-    .enableExperimentalFeature("StrictConcurrency=complete"),
+    .enableUpcomingFeature("InternalImportsByDefault"),
+    .enableUpcomingFeature("MemberImportVisibility"),
+    .enableUpcomingFeature("InferIsolatedConformances"),
+    .enableUpcomingFeature("NonisolatedNonsendingByDefault"),
+    .enableUpcomingFeature("ImmutableWeakCaptures"),
```
MahdiBM, line 46: Is this even available in 6.2? The proposal doesn't mention that but it could be.
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0481-weak-let.md

MahdiBM (approved):
Looks good.

Package.swift
gwynne: No, but upcoming/experimental feature flags that aren't available yet don't fail the build.

Tests/FluentTests/OperatorTests.swift
```diff
@@ -1,10 +1,13 @@
...
-import XCTVapor
-import FluentKit
+import VaporTesting
 
-final class OperatorTests: XCTestCase {
-    func testCustomOperators() throws {
+@Suite
+struct OperatorTests {
+    @Test
+    func customOperators() throws {
```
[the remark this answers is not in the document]
gwynne: My guess is Tanner included this to make sure the operators remain syntactically valid in context. That's a very, very, very valid test, even if it looks like it does nothing. Trust me, it's easy as hell to break them...

